PR #13, opened 2025-07-09
Vrs upload gui

rpuns:
1. added vrs upload gui
2. added task map update and metadata update


rpuns:
> * Can you make the vrsplayer part an optional flag?
> * wrap your upload_vrs_to_s3.py with a main function and if **name** == "main" logic.

will add modifications in next pr shortly

PR #17, opened 2025-07-11
Flow Matching and Diffusion PR

rpuns:
- Added FM and Diffusion Policy
- Added abstract Denoiser Policy class
- Added relevant configs
- minor modifications to support diffusion to hpt.py
- added flexible norm modes (minmax and quantile as well as warnings when using zscore for diffusion)

Should merge directly on top of egobridge pr


rpuns:
Found a bug regarding ac keys

PR #22, opened 2025-07-13
Flow Matching, Diffusion PR

rpuns:
- added UNet Diffusion and Flow Matching
- added cross attention flow transformer
- cleaned up hard coded losses
- added and tested relevant configs

### DIFF
--- a/egomimic/hydra_configs/data/laundry_cotrain.yaml
+++ b/egomimic/hydra_configs/data/laundry_cotrain.yaml
@@ -8,7 +8,7 @@ train_datasets:
         repo_id: "blue"
         root: "/coc/cedarp-dxu345-0/datasets/RLDB/NEW_DATA/aria_eve_laundry_0608/lerobot_6dof"
         local_files_only: true
-        mode: "total"
+        mode: "train"
     embodiment: "eve_bimanual"
   dataset2:
     _target_: rldb.utils.MultiRLDBDataset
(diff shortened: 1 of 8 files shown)


egomimic/algo/hpt.py
```diff
@@ -116,7 +118,14 @@ def __init__(
         self.auxiliary_ac_keys = None
         self.shared_action = False
         self.device = None
-
+        
+        self.ot_6dof = False
+        self.use_dtw = False
```
ElmoPA, line 123: If this is for the ot then mayb make it a mode so that we can flexibly have it so that we can use different pairing methods like L2 and Language dist.

egomimic/algo/hpt.py
```diff
@@ -271,7 +280,7 @@ def preprocess_tokens(self, domain, features):
 
         if self.token_postprocessing == "action_token":
             action_tokens = self.action_tokens.repeat(len(tokens), 1, 1)
-            tokens = torch.cat([tokens, action_tokens], dim=-2)
+            tokens = torch.cat([action_tokens, tokens], dim=-2)
```
ElmoPA, line 283: I will prefer to have action tokens later seems more like intuitively make sense (just feel that inputs is left and outputs is to the right).

egomimic/algo/hpt.py
```diff
@@ -377,7 +386,23 @@ def stem_process(self, domain, data):
             feat_dict[modality] = stem_token
 
         return feats, feat_dict
+    
+    def resume_from_depth(self, block_outputs, depth):
+        """
+        Detach at trunk depth and resume trunk forward pass.
+        Gradients will only flow from depth upward.
+        """
+        cut_tokens = block_outputs[depth - 1].detach()
```
ElmoPA, line 395: What are we using stop gradient for? Also isn't it kinda inefficient to already compute the entire trunk but only use the one at that depth?

egomimic/algo/hpt.py
rpuns: that’s a problem because then the action tokens for each embodiment don’t align - for example robot has additional wrist image which would mean by the aftention matrix, action tokens for human go in the same spot as wrist for robot

egomimic/algo/hpt.py
rpuns: this is for the freeze at depth part for coat / ot loss

egomimic/algo/hpt.py
rpuns: it is a mode but language pairing etc is not relevant for egoverse - will add for egobridge release

PR #47, opened 2025-10-27
Add support for 21 finger keypoints tracking

RogerQi:
- Bump projectaria tool version to 1.6.0.
- Added support for fingertip coordinate tracking
- Tested that fingertip masking and wrist masking are the same on 1.6.0 SDK (wrist avaialble <=> finger keypoints available)


RogerQi:
I've tested `projectaria-tools==1.6.0` and `projectaria-tools==1.5.7` can behave slightly different. For example, the sum of bad pose data in `clean_data` function given by this line

```python
np.any(pose >= 1e8, axis=1)
# I set a breakpoint to get sum using the following line
np.any(pose >= 1e8, axis=1).sum()
```

For VRS file ``, this sum evaluates `3360` on 1.6.0. For the old 1.5.7, the same line evaluates to `3252`.

I've checked the aligned actions match perfectly. So this should have minimum effect.

PR #207, opened 2026-02-18
4 potential bugs

Aseemrd:


egomimic/rldb/zarr/zarr_dataset_multi.py
```diff
@@ -151,13 +151,13 @@ def __init__(
...
+        filters: dict | None = None,
+    ) -> dict[str, "ZarrDataset"]:
         """
-        Outputs a list of ZarrDatasets with relevant filters.
-        If sync_from_s3 is True, sync S3 paths to local_root before indexing.
-        If not True, assumes folders already exist locally.
+        Outputs a dict of ZarrDatasets with relevant filters.
+        Syncs S3 paths to local_root before indexing.
         """
+        filters = dict(filters) if filters is not None else {}
```
Aseemrd, line 160: **Bug: mutable default `filters={}` → `None` + defensive copy.**

The old default `filters={}` was a single shared dict object. When `resolve()` did `filters["is_deleted"] = False`, it permanently mutated that shared default. On the second call without explicit filters, `filters` would already contain `{"is_deleted": False}` plus whatever was added previously — silently corrupting the query.

Now we default to `None` and create a fresh `dict()` copy on every call. Also fixed the return type annotation: this returns `dict[str, ZarrDataset]`, not `list[tuple[str, str]]`.

egomimic/rldb/zarr/zarr_dataset_multi.py
```diff
@@ -201,6 +201,7 @@ def _get_filtered_paths(filters: dict = {}) -> list[tuple[str, str]]:
             list[tuple[str, str]]: List of tuples, each containing (processed_path, episode_hash)
                                    for episodes passing the filter criteria.
         """
+        filters = dict(filters) if filters is not None else {}
```
Aseemrd, line 204: **Same mutable default fix for `_get_filtered_paths`.**

Same pattern — `filters={}` default was shared and could accumulate state across calls. Switched to `None` with defensive copy.

egomimic/rldb/zarr/zarr_dataset_multi.py
```diff
@@ -338,7 +339,11 @@ def _local_filters_match(metadata: dict, episode_hash: str, filters: dict) -> bo
                 continue
 
             if key == "robot_name":
-                meta_value = metadata.get("robot_name", metadata.get("robot_type"))
+                meta_value = (
+                    metadata.get("robot_name")
+                    or metadata.get("robot_type")
+                    or metadata.get("embodiment")
```
Aseemrd, line 345: **Bug: filter matcher now checks `embodiment` metadata key.**

`ZarrWriter._build_metadata()` stores the robot type under the key `embodiment`, but this filter matcher only checked `robot_name` and `robot_type`. Any episode written with the current writer would silently fail to match a `robot_name` filter because the metadata key it was looking for didn't exist.

Added `embodiment` as a fallback so it works with both old and new episode formats.

egomimic/rldb/zarr/zarr_dataset_multi.py
```diff
@@ -457,8 +455,13 @@ def __init__(self,
         else:
             raise ValueError(f"Unknown mode: {mode}")
 
-        datasets = {rid: ds for rid, ds in datasets.items() if rid in chosen}
-        assert datasets, "No datasets left after applying mode split."
+        self.datasets = {rid: ds for rid, ds in datasets.items() if rid in chosen}
```
Aseemrd, line 458: **Bug: MultiDataset split was broken.**

Previously, `self.datasets = datasets` and `self.index_map` were built from ALL datasets before the split logic ran. The split logic only created a local `datasets` variable — `self.datasets` and `self.index_map` were never updated. This meant train/valid splits were silently ignored and all modes returned the full dataset.

Fixed by moving the split logic first, assigning to `self.datasets` (not a local), then building `index_map` from the filtered set.

egomimic/rldb/zarr/zarr_dataset_multi.py
```diff
@@ -491,10 +494,13 @@ def _from_resolver(cls, resolver: EpisodeResolver, **kwargs):
         sync_from_s3 = kwargs.pop("sync_from_s3", False)
         filters = kwargs.pop("filters", {}) or {}
 
-        resolved = resolver.resolve(
-            sync_from_s3=sync_from_s3,
-            filters=filters,
-        )
+        if isinstance(resolver, LocalEpisodeResolver):
```
Aseemrd, line 497: **Bug: `_from_resolver` crashed with `S3EpisodeResolver`.**

The old code unconditionally passed `sync_from_s3` to `resolver.resolve()`, but `S3EpisodeResolver.resolve()` doesn't accept that parameter (it always syncs from S3 internally). This caused a `TypeError` on every call with an S3 resolver.

Fixed with an `isinstance` check — only `LocalEpisodeResolver` gets the `sync_from_s3` kwarg.

PR #452, opened 2026-05-15
HNet variants: delta actions, cross-attn, fused tokens, bigger AdaLN

ElmoPA:
HNet variants: delta actions, cross-attn, fused tokens, bigger AdaLN

Four new variants for ablating the H-Net conditioning architecture on
pushshapes packed training:

1. **Delta actions** (data-side):
   - New ``DeltaAction`` transform that converts absolute action targets
     to per-step deltas (``out[t] = arr[t] - arr[t-1]``, ``out[0] = 0``).
   - New ``data/tsimulation_delta.yaml`` wiring the transform onto the
     packed pushshapes dataset.

2. **Cross-attention** (block-side):
   - New ``CrossMultiHeadAttention`` in ``blocks.py`` — queries from x,
     keys/values from cond tokens. Padded + packed forward + step (with
     its own KV cache that accumulates cond tokens across AR steps).
   - ``TransformerBlock.cond_mode`` selects "adaln" (default) or
     "cross_attn"; the cross-attn variant inserts an extra cross-attn
     layer between self-attn and MLP.
   - ``Isotropic`` threads ``cond_mode`` through to each block;
     ``build_isotropic`` reads it from the spec dict.
   - New ``model/hnet_pushshapes_crossattn.yaml``.

3. **Fused tokens** (policy-side):
   - New ``FlatFusedPolicy`` + ``HNetFused`` algo in ``algo/hnet.py``.
     Bypasses the H-Net stage hierarchy; instead, builds a single causal
     transformer stack and feeds an interleaved sequence
     ``[c_0, BOS, c_1, a_0, c_2, a_1, ...]`` of length 2T. Action
     predictions are extracted from odd positions.
   - Supports forward (padded + packed) and AR generate.
   - New ``model/hnet_pushshapes_fused.yaml``.

4. **Bigger AdaLN** (capacity-only):
   - New ``model/hnet_pushshapes_big.yaml``: same baseline arch but
     d_model 128->256, d_intermediate doubled, num_heads 4->8.

All implementations pass CPU instantiation + forward/generate/packed
smoke tests; the existing 57 hnet_nets unit tests still pass.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

HNet eval: switch to teacher-forced single-pass + 2 recipe variants

**Eval change:** ``HNet.forward_eval`` now runs a single teacher-forced
forward pass (via ``forward_packed`` -> unpack to padded) instead of the
per-episode AR rollout. AR rollout was meaningless for this eval setup:
the obs sequence is fixed (recorded), so the predicted action doesn't
change what the model sees next — AR just compounded exposure-bias
error without testing anything useful. The previous ``_ar_rollout_packed``
method is preserved (dead code for now) for potential closed-loop sim
eval later.

**Two new opt-in variants:**

1. ``model/hnet_pushshapes_recipe.yaml`` — baseline AdaLN + the H-Net
   paper's 3 training tricks (residual-stream-aware init,
   per-stage LR multipliers [3.0, 1.7, 0.9], parameter_groups with
   WD=0 on bias/norm + WD=0.01 elsewhere) + LR 5e-5 (half of baseline).

2. ``model/hnet_pushshapes_fused_lowlr.yaml`` — fused tokens (current
   best) with LR 5e-5 to test whether ep 49->99 val drift was
   overfitting from too-aggressive LR.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

pl_model: fix optimizer instantiation for param-group case

When ``self.model.parameter_groups(base_lr)`` returns a ``list[dict]``,
passing it as a kwarg through ``hydra.utils.instantiate`` wraps the
dicts as OmegaConf ``DictConfig`` objects, which trips AdamW's
tensor-type validation:

  TypeError: optimizer can only optimize Tensors, but one of the params
  is omegaconf.dictconfig.DictConfig

Fix: instantiate the optimizer's ``_partial_`` first, then call it with
the native-Python params arg.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


ElmoPA:
* **#477** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/477?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#452** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/452?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/452?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #563, opened 2026-08-08
fix(robot): import robot_utils by its package path

ElmoPA:
rollout.py and collect_demo.py did:

    from robot_utils import RateLoop

but there is no top-level robot_utils package -- the file is
egomimic/robot/robot_utils.py. That bare form only resolves when the script is
run from inside egomimic/robot/ (that directory on sys.path), so importing
either module as part of the package raised ModuleNotFoundError. It is also why
rollout.py looked dead to reference sweeps: it is the sole caller of
egomimicUtils.interpolate_arr / interpolate_arr_euler, and an import-based check
never got far enough to see it.

stream_aria.py and stream_d405.py already use the package-qualified form; this
makes all four consistent.

Verified: egomimic.robot.rollout now imports (previously ModuleNotFoundError).
collect_demo still fails on ppadb, an unrelated optional hardware dependency for
the Oculus reader that is simply absent from the cluster venv.

### DIFF
--- a/egomimic/robot/collect_demo.py
+++ b/egomimic/robot/collect_demo.py
@@ -22,7 +22,7 @@ sys.path.append(os.path.join(os.path.dirname(__file__), "oculus_reader"))
 from oculus_reader import OculusReader
 
 # Import local modules
-from robot_utils import RateLoop
+from egomimic.robot.robot_utils import RateLoop
 
 # Add path to robot_interface
 sys.path.append(os.path.join(os.path.dirname(__file__), "eva/eva_ws/src/eva"))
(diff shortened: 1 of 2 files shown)


ElmoPA:
* **#563** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/563?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #562, opened 2026-08-08
chore: remove seven unused modules, rename a mis-named converter

ElmoPA:
chore: remove seven unused modules, rename a mis-named converter

Deletions -- each has zero importers on main, verified by module-path search,
per-symbol search, hydra _target_ strings, `python -m` invocations, file-path
references in shell/CI, and dynamic-import patterns:

  egomimic/utils/memory_utils.py            229
  egomimic/utils/real_utils.py              142
  egomimic/pl_utils/test_model_wrapper.py   135
  egomimic/rldb/compression_utils.py        130
  egomimic/rldb/data_utils.py               103
  egomimic/rldb/scripts/utils.py             86
  egomimic/utils/obs_utils.py                10

Two worth calling out:

  * rldb/scripts/utils.py is a dead DUPLICATE -- all five of its functions
    (nds_pq, is_key, is_listy, nds, str2bool) also exist in
    egomimic/utils/egomimicUtils.py, and every caller imports them from there.

  * compression_utils.py is the superseded video codec (encode_video /
    decode_video on PyAV). Its replacement is rldb/zarr/video_codec.py, which
    arrives with the h264 PR. Nothing on main imports compression_utils. The
    only importer anywhere is rldb/zarr/verify_zarr.py, which exists solely on
    six branches last touched in February 2026 and on neither main nor the
    batchflow stack; those branches carry their own copy of that file and would
    need a rebase regardless.

Rename:

  egomimic/rldb/zarr/zarr_write_test.py -> egomimic/rldb/zarr/hdf5_to_zarr.py

The name said "test", but it is not one: it is an HDF5 -> Zarr v3 conversion
utility with an argparse CLI (--hdf5-path / --arm / --extrinsics-key), forward
kinematics and camera-frame transforms. The new name matches the existing
aria_to_zarr / eva_to_zarr / mecka_to_zarr convention. Nothing referenced the
old name.

Deliberately KEPT, having been checked rather than assumed:

  * benchmark_forward_pass.py -- a working dataloader-throughput benchmark with
    an argparse CLI and no hardcoded paths.
  * test_zarr.py -- a genuine pytest module (13 tests over MultiDataset /
    ZarrDataset / action-chunk transforms). It is currently uncollected because
    of five hardcoded absolute paths and an import of
    build_human_bimanual_transform_list, which exists on no branch. That makes
    it a broken test to repair, not dead code to delete.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

[...]

PR #565, opened 2026-08-09
fix(zarr): repair an unresolvable import, and share two copy-pasted helpers

ElmoPA:
egomimic/rldb/zarr/hdf5_to_zarr.py imported EvaHD5Extractor from
egomimic.scripts.eva_process.zarr_utils, which exists on no branch -- the class
lives in eva_utils, which the sibling eva_to_zarr.py:18 imports correctly. The
module was therefore unimportable and its CLI could not run. Predates the rename
in #562 (that commit changed no content).

safe_rot3_from_T was defined identically in robot/calibrate_utils.py and
robot/collect_demo.py. It moves to utils/pose_utils.py, next to the other pose
maths. Deliberately NOT hoisted into either robot module: calibrate_utils does a
sys.path.append + "from oculus_reader import OculusReader" at import time, so
importing it for one numpy helper would drag a VR hardware dependency into
collect_demo. pose_utils has no such deps.

The four single-arm from32 converters each re-implemented the same 10-wide block
decode, differing only in offset (left 0, right 10) and whether the gripper
channel is kept -- robot returns 7-dim, human 6-dim. The decode, including the
rotation reconstruction, moves to _decode_arm_block; each converter keeps its own
cat so the arity difference stays visible.

Verified by differential test: all four from32 are bit-identical to the previous
implementation over random inputs. (The ~pi round-trip error on random ypr is a
pre-existing _matrix_to_ypr wrap-around, identical before and after, and vanishes
for ypr in the principal range.)

### DIFF
--- a/egomimic/rldb/zarr/hdf5_to_zarr.py
+++ b/egomimic/rldb/zarr/hdf5_to_zarr.py
@@ -21,7 +21,7 @@ from pathlib import Path
 import numpy as np
 
 from egomimic.rldb.zarr import ZarrWriter
-from egomimic.scripts.eva_process.zarr_utils import EvaHD5Extractor
+from egomimic.scripts.eva_process.eva_utils import EvaHD5Extractor
 from egomimic.rldb.embodiment.eva import Eva
 
 
--- a/egomimic/robot/calibrate_utils.py
+++ b/egomimic/robot/calibrate_utils.py
@@ -5,6 +5,7 @@ import time
 
 import numpy as np
 from scipy.spatial.transform import Rotation as R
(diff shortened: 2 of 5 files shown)


ElmoPA:
* **#565** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #610, opened 2026-09-08
perf(arc): vectorize arc-length resampling, 28x faster tokenize

aidang3019:
perf(arc): vectorize arc-length resampling, 28x faster tokenize

resample_by_distance and the deploy-time detokenize both interpolated one
arc-length target at a time, and _interp_ypr_at_s builds a fresh two-keyframe
scipy Slerp per call. A bimanual token at M = 100 is ~200 Rotation/Slerp
constructions per sample, which dominates data loading: 33.5 ms/sample on real
ABC chunks (200 frames, D = 0.40 m, M = 100), and ~1 s/sample inside loader
workers that inherit a full BLAS thread pool, where it starved the GPU (first
arc smokes ran at 8.7 samples/s, 25 min per 100-batch epoch).

resample_at_arc_lengths does the whole target vector in one pass: one
searchsorted to bracket every target, one linear blend for position and
gripper, and one batched r0 * exp(alpha * log(r0^-1 r1)) for rotation -- the
same geodesic scipy walks for two keyframes. Same call sites, same semantics,
same clamping at both ends. 1.18 ms/sample, 28x.

Equivalence is exact, not approximate: over 96 real arm-tracks the vectorized
resample matches the per-target helpers to 0.00e+00 on position, rotation and
gripper. The scalar helpers stay in the module and are now the test's
reference implementation.

Claude-Session: https://claude.ai/code/session_01B8hXrMvNvrdEdK4MCYoNLp

feat(arc): velocity_norm option, so the velocity row is a speed and not a shape

MEAN_PER_DIM stores (pos[-1] - pos[0]) / total_time -- a CHORD rate. df5b8498
made the decoder correct by converting it to the arc rate it needs (x total /
chord), so round-trip timing is already right and this changes none of it:
measured on 320 real ABC-skirts arm-tokens, chord and path replay to the same
duration (median 0.984 of the truth, q10 0.957, q90 1.000, identical to three
decimals under both norms).

What the conversion cannot fix is what the row IS. chord / total_time
conflates how fast the arm moved with how much its path doubled back. On real
fold tokens at D = 0.40 m, chord / path spans 0.037 to 0.977 -- a 26x
multiplicative confound (53x at D = 0.75). Two motions at the same speed can
differ 26x in the value the head has to regress, and the head does have to
regress it: the velocity row is part of the predicted token, not metadata.
A path-normed row is the speed alone, so its spread across a task is the
tempo spread and nothing else.

velocity_norm="path" keeps the chord direction and swaps the magnitude to
path / total_time. Duration then recovers as path / speed from the waypoints'
own polyline length, which is never ~0 for a token that moved -- unlike
chord / speed, which is undefined when a path returns to where it started and
today silently falls back to (M-1)*dt. That case is real but rare: no token in
~880 sampled here had chord / path below 0.01, so treat it as robustness
rather than a bug being hit.

Default is "chord" -- existing configs and checkpoints are untouched, and a
test asserts the default token is byte-identical to the pre-change one.

Claude-Session: https://claude.ai/code/session_01B8hXrMvNvrdEdK4MCYoNLp

### DIFF
--- a/egomimic/rldb/zarr/arc_length_tokenizer.py
+++ b/egomimic/rldb/zarr/arc_length_tokenizer.py
@@ -199,6 +199,87 @@ def _interp_linear_at_s(
     return (1.0 - alpha) * values[i] + alpha * values[i + 1]
 
 
+def _bracket_segments(
+    cumdist: np.ndarray, targets: np.ndarray
+) -> tuple[np.ndarray, np.ndarray]:
+    """Vectorized :func:`_bracket_segment` over many arc-length targets.
+
+    Same clamping rules as the scalar version: a target at or below
+    ``cumdist[0]`` maps to ``(0, 0.0)``, at or above ``cumdist[-1]`` to
+    ``(len - 2, 1.0)``, and a zero-length segment yields ``alpha = 0``.
+    """
+    cumdist = np.asarray(cumdist, dtype=np.float64)
+    targets = np.asarray(targets, dtype=np.float64)
+    n = len(cumdist)
+    last = max(n - 2, 0)
+    i = np.clip(np.searchsorted(cumdist, targets, side="left") - 1, 0, last)
+    s0 = cumdist[i]
+    s1 = cumdist[np.minimum(i + 1, n - 1)]
+    span = s1 - s0
+    ok = span > 1e-12
+    alpha = np.where(ok, (targets - s0) / np.where(ok, span, 1.0), 0.0)
+    lo = targets <= cumdist[0]
+    hi = targets >= cumdist[-1]
+    i = np.where(lo, 0, np.where(hi, last, i))
+    alpha = np.where(lo, 0.0, np.where(hi, 1.0, alpha))
+    return i, alpha
+
+
+def _slerp_segments_ypr(
+    ypr: np.ndarray, i: np.ndarray, alpha: np.ndarray
+) -> np.ndarray:
+    """SLERP ``ypr[i] -> ypr[i + 1]`` at ``alpha``, for many pairs at once.
+
+    Walks the same geodesic scipy's two-keyframe ``Slerp`` walks, as one
+    batched rotation-vector step instead of one ``Slerp`` object per target.
+    """
+    ypr = np.asarray(ypr, dtype=np.float64)
+    n = len(ypr)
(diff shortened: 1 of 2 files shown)


aidang3019:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/610?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#610** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/610?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#573** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/573?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#611](https://github.com/GaTech-RL2/EgoVerse/pull/611) <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/611?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#606** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/606?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#612** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/612?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#605** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/605?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#604** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/604?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#176](https://github.com/GaTech-RL2/EgoVerse/pull/176) <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/176?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

PR #616, opened 2026-09-10
Fix PI0.5 raw rotation action encoding

aidang3019:
Fix PI0.5 raw rotation action encoding

Preserve physical YPR angles when converting normalized Cartesian actions to 6D rotation targets so PI training does not learn rotations from normalized Euler coordinates.

Co-authored-by: Cursor <cursoragent@cursor.com>
(cherry picked from commit 349e6c804bea6b7606376fad366abacaf0797a98)

Tighten raw rotation action encoding

Keep the new PI0.5 encoding simple by dropping the version suffix and failing fast when a converter does not implement raw-rotation packing.

Co-authored-by: Cursor <cursoragent@cursor.com>
(cherry picked from commit b2223eaa6788a762598ff4561f0fa7b25e5915d9)

Harden raw rotation action handling

Clone normalized actions before unnormalization and assert the decoded bimanual action shape before applying non-rotation unnormalization.

Co-authored-by: Cursor <cursoragent@cursor.com>
(cherry picked from commit 309e5f99385d2a90c9962ca1191d08568d7cf948)

6D pi data transform

(cherry picked from commit 6494762f2d22c534565422aaa1c6f990676c9312)

6D proprio, rotation-free bounds check, constant LR, mecka 6D configs

- eva/human cartesian_6d + wristframe_6d modes 6D-encode
  observations.state.ee_pose (revert lists convert it back before the
  eef-frame revert) so proprio normalization avoids the ±π yaw/roll wrap
- bounds checking skips rotation channels of actions_cartesian / ee_pose
  via BIMANUAL_CARTESIAN_LAYOUTS (ported from the wristframe-6d stack)
- pi0.5_base: constant-with-warmup LR schedule (cosine decayed to 0 @ 60k)
- new pi0.5_bc_mecka_6d + mecka_all_pi_6d configs: all SQL mecka episodes,
  bs 64, cartesian_6d stride 1, proprio in prompt, no embodiment label

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
(cherry picked from commit 19c59e50f0e3eb920288c140719dee72e3d81546)

mecka wristframe-6d: launch infra, precomputed norm stats, left-wrist Rz(180) fix

Squash of the two mecka-all pipeline commits that used to sit inside
aidan/eval_gating, so the whole wristframe 6D pipeline lives on this
branch.

mecka-all launch infra (was a8ae76b6):
- pi0.5_base default_prompt "" (was "This is a bad action."; remote parity
[...]

PR #617, opened 2026-09-10
Infra fixes for training

RyanPCo:
# Changes on this branch (status: implemented 2026-09-10)

Everything below landed in code and configs. What did NOT land is the part
that has to run on PACE, listed under "Still to do" at the end.

## Only one PI0.5 action encoding: `cartesian_normalized_rot6d` (done)

**Decision:** keep `cartesian_normalized_rot6d`. Delete both
`legacy_normalized_ypr_rot6d` and `cartesian_ypr_raw_rot6d`.

### The three modes before this change

| Mode | Where yaw/pitch/roll → 6D happened | Status |
|---|---|---|
| `legacy_normalized_ypr_rot6d` | In the model, on *normalized* numbers (bug) | deleted |
| `cartesian_ypr_raw_rot6d` | In the model: un-normalize the rotation back to radians, then convert | deleted |
| `cartesian_normalized_rot6d` | In the data loader (`CartesianYPRToRot6D`), before normalization; stats are computed on 6D | **the only path** |

### Why this one

- **Legacy was wrong.** It converted *normalized* yaw/pitch/roll to 6D as if
  they were angles. The targets weren't real rotations and they jumped at the
  ±π yaw wrap (`test_ypr_to_rot6d_transform_is_continuous_at_yaw_wrap` now
  pins the continuity of the data transform instead).
- **Raw-rot6d was correct but put rotation logic in the model.** It had to
  un-normalize, split rotation dims from the others, and do its own partial
  normalization inside `pi.py` and the converters.
- **Normalized-rot6d keeps the model simple.** The data is already 6D and
  normalized when it arrives, so the model only packs it to 32-D. Rotation math
  lives in one place (the data transform), and proprio (`ee_pose`) gets the same
  6D treatment there. The cost is that norm stats must be computed on the 6D
  data.

### What was removed

`egomimic/utils/action_utils.py`
- The three encoding constants, `ROBOT_BIMANUAL_CARTESIAN_*_DIMS`,
  `_normalize/_unnormalize_robot_bimanual_non_rot`.
- Every `to32` / `from32` / `to20` / `from20` and `*_raw_rotation` method.
  With one encoding the only pack/unpack is `to32_norm_6d` /
  `from32_norm_6d` (base class raises, the two bimanual converters implement).
- The four single-arm converter classes (no yaml named them).
- `HumanBimanualKeypoints` (keypoint path, below).
- Kept on purpose: `_stat_tensor`, `_apply_norm_one`, `_apply_unnorm_one`
  (tensor twins of the dataset normalization, used by
  `test_wrist6d_roundtrip.py`) and the rotation helpers `_ypr_to_matrix`,
  `_matrix_to_ypr`, `_reconstruct_R_from_cols` (used by the eval metrics and
  tests).

`egomimic/algo/pi.py`
- The `action_encoding` constructor arg, attribute and dispatch; `_action_stats`,
  `_unnormalize_action`, `_extract_xyz` (ypr-era, unused).
- `_resize_action_projections` and its call. Replaced by a hard check: the
  vendored openpi port hard-codes 32-wide action projections, so
  `model.action_dim != 32` now raises instead of silently mismatching.

`egomimic/eval/eval_latent.py`: `converter.from32()` → `from32_norm_6d()`.

Tests: `test_pi05_action_encoding.py` deleted; the yaw-wrap continuity check
moved to `test_pi05_norm_rot6d.py` as a test of `CartesianYPRToRot6D`.

### Config migration (done in-repo; norm stats + retrains are on PACE)

Instead of giving every ypr data config a `_6d` twin, the shared base
`data/cotrain_pi_base.yaml` now sets `transform_list.mode: cartesian_6d` for
both embodiments, so every config that inherits it is 6D: `eva_pi`,
`eva_pi_lang`, `cotrain_pi_lang`, `cotrain_pi_no_lang`, `obj_gen_pi_lang`,
`cotrain_pi_latent`. Explicit flips: `cotrain_pi_lang_wrist` and
`obj_gen_pi_lang_wristframe` → `cartesian_wristframe_6d`;
`cotrain_pi_pickplace_qwen` → `cartesian_6d`.

- Deleted `data/cotrain_pi_lang_6d.yaml` (identical to `cotrain_pi_lang` now).
  `obj_gen_pi_lang_6d` and `obj_gen_pi_lang_wristframe_6d` stay: they also
  switch to the AnnotationCutoff resolver, which is the real difference now.
- Deleted `evaluator/eval_pi.yaml` (the ypr revert; nothing produces ypr).
  `train_zarr_cartesian_pi.yaml` defaults to `evaluator=eval_pi_camframe_6d`.
- `action_encoding:` removed from `pi0.5_bc_eva`, `pi0.5_bc_mecka_6d`,
  `pi0.5_cotrain_eva_aria_6d`. The "legacy (default)" model yamls
  (`pi0.5_bc_aria`, `pi0.5_bc_mecka`, `pi0.5_bc_scale`,
  `pi0.5_cotrain_mecka_scale`, `pi0.5_cotrain_eva_aria`) needed no change: the
  model no longer has an encoding knob, the data config decides.
- Verified: every PI data/model/evaluator combination composes through
  `train_zarr_cartesian_pi`, every resolved `transform_list.mode` ends in
  `6d`, and the transform lists instantiate.

**Old checkpoints: no compatibility path.** Checkpoints trained with legacy
or raw-rot6d won't decode after this change. Anything still needed gets
evaluated from the pre-deletion commit (`2c091eb8`).

## Trim redundant validation metrics (`egomimic/eval/eval_pi.py`) (done)

**Decision:** log fewer, easier-to-read metrics. Most of the `cam_*` metrics
repeated a native metric exactly.

### Why the cam metrics were redundant

The evaluator's revert to camera frame applies the **same** rigid transform
(the observed wrist pose from the batch) to both prediction and ground truth:
`cam = T_wrist @ chunk`. A shared rotation and translation doesn't change
distances or angles:

- Position: `‖R_w (p − g)‖ = ‖p − g‖`, so `cam_xyz_paired_mse_avg == xyz_paired_mse_avg`.
- Rotation: `(R_w R_p)ᵀ (R_w R_g) = R_pᵀ R_g`, so the cam-frame rotation error equals the native one.

This holds for both wrist-frame 6D configs (frame change + format change) and
camera-frame 6D configs (format change only), and for DTW / Fréchet on xyz.
`test_pi_eval_metrics_are_invariant_to_a_shared_rigid_transform` pins it.

### Logged now (per embodiment, native 6D output)

| Metric | Meaning |
|---|---|
| `xyz_paired_mse_avg` | Position error (m²) over the chunk |
| `xyz_final_mse_avg` | Position error at the last timestep only (longest horizon, hardest) |
| `rot_err_deg_avg` | Rotation error in **degrees**: geodesic angle `arccos((tr(R_pᵀR_g)−1)/2)` between the Gram-Schmidt-decoded prediction and gt. No ±π wrap, no gimbal lock, axis-independent. |
| `rot_err_deg_final` | Same, last timestep only |
| `rot6d_paired_mse_avg` (secondary) | MSE over the raw 6D columns, the channels the flow-matching loss regresses (up to per-dim normalization). Pre-Gram-Schmidt, axis-dependent, unitless: a training-tracking diagnostic, not the headline. Was mislabeled `ypr_paired_mse_avg`. |
| `xyz_dtw_avg` | DTW on the xyz channels only (metres, per-step normalized). Forgives temporal misalignment. |
| `xyz_frechet_gauss_avg` | Gaussian Fréchet (2-Wasserstein) distance between the time-distributions of prediction and gt, xyz only. Shape of the motion rather than step-by-step alignment. |
| `bestof{M}_paired_mse`, `sample_diversity_M{M}` | Only when `val_num_samples` > 1. Multimodal coverage and spread from M stochastic draws. |
| `action_loss`, `<embodiment>_loss` | Training objective |

### Removed

`cam_paired_mse_avg`, `cam_final_mse_avg`, `cam_*_nowrap`,
`cam_xyz_paired_mse_avg`, `cam_rot_geodesic_avg`, `cam_ypr_paired_mse_avg`
(cam path duplicates or mixed-unit / Euler-angle MSE); native `paired_mse_avg`,
`final_mse_avg` (mixed units); `rot_geodesic_avg` (renamed, now degrees);
`reverse_kl_M{M}`, `mean{M}_paired_mse`, `worstof{M}_paired_mse`;
`dtw_avg`, `frechet_gauss_{avg,min,max}` on the full mixed-unit vector
(replaced by the xyz-only versions). `_wrap_aware_mse` and
`reverse_kl_from_samples` are gone with them.

### Follow-ups (done)

- Sample-count knob renamed: `config.reverse_kl_samples` →
  `config.val_num_samples`, `algo.rkl_samples` → `algo.val_samples`
  (`pi.py`, `eval_pi.py`, `eval_train_viz.py`, `pi0.5_base.yaml`,
  `pi0.5_bc_mecka_6d.yaml`, `eval_cart_ep.sbatch`, tests).
- `eval_hpt.py`: the reverse-KL block and `_collect_policy_samples` are gone,
  as is `rkl_samples` in `hpt.py` and the `reverse_kl_samples:` line in the
  nine HPT model yamls (it was read from `**kwargs`, so nothing else changes).
  HPT keeps its own paired/final/cam/Fréchet metrics: it still runs main's
  keypoint action spaces, where the xyz-only split does not apply.
- Cam revert only runs when rendering video: `PIEvalVideo.compute_metrics_and_viz`
  skips `Embodiment.apply_transform` (and the batch deepcopy) unless `do_viz`,
  i.e. on every validation batch that isn't rendered (`viz_every_n_epochs` /
  `viz_max_batches`).

### Naming: say `6d` when it's 6D, `ypr` only when it's yaw/pitch/roll

Use `ypr` in metric keys, variable names, and docstrings only for values that
really are yaw/pitch/roll. Use `rot6d` for 6D rotation-matrix columns. Use
`rot` for metrics that don't depend on the format (`rot_err_deg_avg`).
Rotation error is reported in degrees; the key carries the unit (`deg`), the
docstring says it is the geodesic angle.

Rotation format at each point of validation:

| Where | Rotation format | Why |
|---|---|---|
| Native prediction / ground truth (`forward_eval` output, unnormalized batch) | **6D** (18-D human / 20-D robot) | The model's native format; all metrics read this |
| Viz path (evaluator `transform_lists`, `do_viz` only) | **ypr** (12/14-D) | `CartesianRot6DToYPR` runs first, because the wrist → head frame revert (`ActionChunkCoordinateFrameTransform`, `mode="xyzypr"`) and `viz_gt_preds` both work in xyz+ypr |
| `_rot_geodesic_error` | Either | Builds rotation matrices from whichever format it gets |

Open question (low priority): the viz path only needs ypr because the frame
revert and the overlay work in Euler angles. Switching them to rotation
matrices would remove the last ypr step from validation.

## Remove the fold-freeform fit-test configs and launchers (done)

Deleted `data/mecka_fold_freeform_pi_6d.yaml`, `data/mecka_fold_freeform_pi_cam6d.yaml`,
`scripts/norm_fold_freeform_{wrist6d,cam6d}.sbatch`,
`scripts/train_{wristframe,camframe}_fold_freeform.sbatch` (commit 124027b4;
one-off wrist-vs-cam comparison, hard-coded to one user's PACE paths,
referenced only by each other). Kept `evaluator/train_viz_pi_camframe_6d.yaml`
(the opsplit cam-frame launcher uses it); its header now points at the opsplit
data config.

## Remove the hand-keypoint action path: hand end-effector only (done)

**Decision:** for now we only train on the hand end-effector (the cartesian
wrist pose, `actions_cartesian`). Every keypoint-action piece this branch
added is gone: the PI keypoint recipe (commits f3e99878, 91fdfdef), the
palm-origin keypoint eval (a7945713, 8fbd49b1, 2aac7512) and the HPT keypoint
twin (2c091eb8): 21 files, plus `HumanBimanualKeypoints`, the
`keypoints_wristframe_pi` transform mode, the `include_ee_pose` keymap flag
(the `fix_mecka_left_wrist` guard again rejects every keypoint mode) and the
`action_dim != 32` projection resize in `pi.py`.

If keypoints come back later, the design to use is still: per hand
`[wrist xyz (3) | wrist rot6d (6) | 21 keypoints (63)]` = 144-D on the
normalized-rot6d path, with a layout-table entry so the bounds check and the
rotation-error metric understand it, and a real (not resized) 144-wide action
head.

**Scope:** only what this branch added. The keypoint modes and viz already on
`main` (`keypoints_headframe_*`, `keypoints_wristframe_ypr/quat`, the MANO
keypoint overlay) are untouched.

Leftovers cleaned in kept files: the opsplit data yaml headers no longer name
the keypoint twin; `eval_cart_ep.sbatch` is the generic offline cartesian eval
launcher; the opsplit launcher headers no longer mention the keypoint recipe.
`viz_max_batches` (77751e6d) stays: the cartesian evaluators use it.

## Held-out-operator split: one list in the data yaml (done)

- `mecka_fold_freeform_opsplit_pi_6d.yaml` has a top-level
  `held_out_operators` list (31 sorted ids). The train and valid filter
  lambdas interpolate it (`${.....held_out_operators}`, relative so it works
  under any mount point), so the two sides cannot drift apart.
- `mecka_fold_freeform_opsplit_pi_cam6d.yaml` inherits the wrist-frame file
  via `defaults` and overrides only `transform_list.mode: cartesian_6d`.
- The header records how the 31 were chosen (2026-09-01, seed 42, skip 10
  largest, >= 10 % of frames) and says not to edit the list because
  checkpoints depend on it.
- Verified: both configs compose through `train_zarr_cartesian_pi` and, after
  resolution, are byte-identical to the previous versions apart from the
  set-literal → list-literal change; every old held-out id is rejected by the
  train filter and accepted by the valid filter, and unknown ids do the
  reverse. With the keypoint configs gone the list exists exactly once.

Upgrade path, if a third consumer of this split appears or a test asserting
the val episode set is wanted: move the list to a `splits/` file with a
`HeldOutFilter(key, values_file, side)` in `egomimic/rldb/filters.py`. Not
needed now. Hashing operator ids or computing the split from SQL at runtime
were rejected (uncontrolled val fraction; val set drifts as data lands).

## Found on the way (pre-existing, not caused by this branch)

- `data/eva_pi.yaml` used an absolute `${train_datasets.eva_bimanual}`
  interpolation, which does not resolve under the `data:` mount, so the plain
  `train_zarr_cartesian_pi` entry point failed to compose. Fixed here
  (relative `${..train_datasets...}`). `aria.yaml`, `aria_pi.yaml`,
  `mecka_pi.yaml` and `aria_qwen.yaml` have the same pattern and were left
  alone (not PI 6D configs).
- `data/cotrain_pi_lang_wrist.yaml` overrode `keymap_mode` to
  `cartesian_wristframe_ypr` / `cartesian`, which Eva's keymap treats as
  "non-PI camera names", so a PI run would have looked for the wrong image
  keys. The overrides are gone; it inherits `cartesian_pi` from its parent.
- The `aria_bimanual` datasets in `obj_gen_pi_lang*.yaml` define no
  `_target_` for the dataset, key_map or transform_list (they do not inherit
  `cotrain_pi_base`, which only knows `eva_bimanual` / `human_bimanual`).
  They compose but cannot instantiate. Left as-is; whoever owns the obj_gen
  experiment should decide whether to fix or delete them.

## Still to do (PACE, not in this repo)

1. **Norm stats.** Every PI data config now emits 6D actions and proprio, so
   any precomputed norm stats for the formerly-ypr configs (`eva_pi`,
   `cotrain_pi_lang`, `obj_gen_pi_lang*`, `cotrain_pi_pickplace_qwen`, ...)
   are stale and the provenance check will reject them. Recompute with
   `scripts/precompute_norm_stats.py` before training. The mecka 6D configs
   (`mecka_all_pi_6d`, the opsplit pair) were already 6D; their stats stand.
2. **Retrains** for anything previously trained on legacy / raw-rot6d
   (`pi0.5_bc_eva` on `eva_pi` first: it is the plain entry point).
3. **Dashboards / wandb panels** that plot the removed keys (`*_paired_mse_avg`,
   `*_final_mse_avg`, `cam_*`, `rot_geodesic_avg`, `reverse_kl_*`, `dtw_avg`,
   `frechet_gauss_*`) need the new names above.
4. **Kept launchers** (`train_*_fold_freeform_opsplit.sbatch`,
   `norm_fold_freeform_opsplit_*.sbatch`, `eval_cart_ep.sbatch`) still default
   `WT` to the `kp-wristframe-port` worktree. Once this branch merges, point
   the default at the primary checkout.


RyanPCo:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/617?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#617** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#616** <a href="https://app.graphite.com/github/pr/GaTech-RL2/EgoVerse/616?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #641, opened 2026-09-15
norm stats: drop samples carrying the 1e9 missing-frame sentinel

RyanPCo:
Aria writes 1e9 pose rows for an undetected hand. Unlike Mecka's zero rows
(scipy rejects a zero quaternion, so those samples are resampled), a 1e9 row
passes the transforms as ~1e9 and reached mean/std/min/max, since
_drop_nonfinite_rows only removed NaN/Inf. It is now _drop_invalid_rows and also
drops any sample with |v| >= schema.POSE_SENTINEL_THRESHOLD (1e8), the same
threshold egoverse validate uses for missing frames. Training still rejects
those samples per item via the bounds check once stats exist.

Tests: tests/unit/test_norm_stats_finite.py adds a unit case and an end-to-end
case on a synthetic Aria episode with 1e9 rows in right.obs_ee_pose (fails
without the fix: observations.state.ee_pose std is poisoned). Full tests/unit:
462 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BbQNNaMAbTNMypegmJkv4s


RyanPCo:
Closing: no longer needed.

PR #662, opened 2026-09-17
ABC-DiT (Amazon FAR) at all four published sizes

RyanPCo:
Adds ABC-DiT ([amazon-far/abc](https://github.com/amazon-far/abc), Apache-2.0) as a selectable model at every size ABC publishes.

**The four sizes are not in the ABC repo.** `abc_minimal/config.py` ships one shape (the xL) and the paper gives only totals — "the total number of parameters in S/B/L/xL are 153M, 290M, 746M, and 1.93B" (Fig. 21). S/B/L are the standard DiT ladder; building ABC's own `DiTPolicy` at each shape reproduces those numbers exactly, which is what pins the table:

| | depth × hidden × heads | + ABC's 85.7M DINOv3 | paper |
| --- | --- | --- | --- |
| S | 12 × 384 × 6 | 153.2M | 153M |
| B | 12 × 768 × 12 | 290.1M | 290M |
| L | 24 × 1024 × 16 | 746.4M | 746M |
| xL | 32 × 1536 × 24 | head alone 1930.0M | 1.93B (head, Table 2) |

- `c80323f2f` **`abc_dit_nets.py`** — the port. A DINOv3 tower whose patch tokens are attention-pooled to 12 query tokens per camera, one 512-d CLIP task vector, and a DiT over the action chunk whose blocks cross-attend to the vision tokens while a 9-way AdaLN modulation carries `[state, task, flow time]`. Module names, shapes and the forward algebra follow the reference, including ABC's state masking and its action-prefix (real-time chunking) sampler. Three things differ, all plumbing: the encoders come from HF (`dinov3-vitb16-pretrain-lvd1689m`, `clip-vit-base-patch32`) rather than vendored loaders reading Meta's `.pth` and a TorchScript CLIP; the images are squash-resized like `SigLIPStem` rather than letterboxed; and actions/state are EgoVerse-shaped — padded shared action width with a loss mask, per-embodiment state embedders, proprio history flattened into the state vector. ABC's frozen `img_proj` is dropped (it exists only so old checkpoints load).
- `9dd3f9e84` **`ABCDiT` algo** — standalone rather than built on the FlowVLA chassis, because ABC conditions differently and carries its own training knobs. Same contract on both ends: HPT-style batches in, the unnormalised-prediction contract `eval_hpt.py` reads out, so the existing evaluators work unchanged. `embodiment_label` defaults off (ABC conditions on the bare task text, and a frozen text tower makes the prompt a read-only channel), and the algo checks the embodiment's cameras against the policy's `camera_roles` — the policy holds one query set and camera embedding per role, in that order, so a silent mismatch would pool the wrong camera through the wrong queries.
- `28967a9a3` **configs** — three levels: `abc_dit_base` (both encoders, ABC's pooling and flow knobs, its AdamW(0.9, 0.95) / wd 0.01 / lr 1e-4 and the paper's 0.1 vision LR scale), `abc_dit_bc_mecka_{kp,6d}` for the action layout, and eight leaves `abc_dit_bc_mecka_{kp,6d}_{s,b,l,xl}` that set the size and nothing else — ABC trains every rung with identical hyperparameters. Plus recipes `train_zarr_mecka_flagship_{kp,6d}_abc_dit`, which take the same data, evaluators and val heads as the hpt / qwenvla / flowvla flagship recipes; they select `b`, the rung closest to the HPT-300M reference arm, and every other rung is a one-word `model=` override.
- `cab011fb3` **tests** — the size ladder re-derived against the paper's counts, the flow-matching convention pinned against a hand-rolled velocity, ABC's state masking and prefix conditioning, the fused-vs-split camera paths, prompts, batch conversion, the deterministic eval pass, all eight configs, and an ABC-DiT arm in `test_train_step` (two optimizer steps through the real `trainHydra.train()` path). Neither pretrained encoder is ever loaded: `stub_abc_encoders` stands in for both by `_target_` swap.
- `595e8cdea` **sampling** — ABC's own 10 Euler steps. The other arms sample with 50 and every val metric is computed from sampled actions, so a number read across arms is not step-matched; re-reading a checkpoint at 50 is an eval-only override.

Only two parameter groups are frozen: the CLIP text tower (63.4M — ABC precomputes task vectors in its dataloader, so they are a fixed feature by construction) and ABC's sincos `pos_embed`. The DINOv3 tower trains at `backbone_lr_scale × lr`, the paper's 0.1 (note `abc_minimal`'s own `OptimConfig` ships 1.0). Plain `trainer=ddp` suffices — unlike FlowVLA, every trainable tensor takes a gradient each step, and a test asserts it.

## Testing

Full unit suite on a compute node: **1159 passed, 2 skipped, 0 failed** (51 new ABC-DiT tests plus the train-step arm).

GPU smoke with the real weights, ABC-DiT-B at the flagship keypoint layout, batch 8 on a TITAN Xp: 275.6M trainable + 63.5M frozen, `action_loss` 2.41 at init falling over four AdamW steps, 0.33 s/step, 5.4 GiB peak; a 50-step Euler sample returns a finite `(8, 100, 144)` chunk in 1.55 s. Trainable parameters at that layout: S 138.9M, B 275.6M, L 731.8M, xL 2000.8M.

Not trained yet — this is the model, not a run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012ER1EbUcuw4hhvV5VZar6h


RyanPCo:
Closing: no longer needed.

